PR #123, opened 2018-06-05
fix prometheus_targets default value.

soloradish:
The `prometheus_targets` default value should be `{}`.


SuperQ:
FYI, to get the test to pass, you also need to update `molecule/default/tests/test_default.py` now that `node.yml` is not populated by default.

PR #132, opened 2018-07-15
Fix custom rule/target file copy

SuperQ:
[patch]
Remove placeholder dirs as they cause Ansible to ignore local files
dirs. Simplify fileglob by only copying valid filenames.


SuperQ:
Fixed up with `[patch]`. :)

PR #151, opened 2018-10-07
Add support for alert relabeling

juliusv:
All other forms of relabeling (targets, metrics, remote write) are
already supported, but alert relabeling wasn't possible yet because it's
not part of the Prometheus Alertmanager configuration.

The most common use case for alert relabeling is dropping the "replica"
label of an HA Prometheus pair, so I chose that as an example.


juliusv:
@SuperQ Didn't know you were an approver in this repo :) Thanks! I still can't merge by myself btw., someone will have to do that for me.

PR #160, opened 2018-11-07
Fix running Dry Mode and improved tag usage

krzyzakp:
- Fixed dry_run (`--check`) mode. By default it was failing on task _unpack prometheus binaries_ as binaries were not downloaded. This changes are done locally, so not danger
- Fixed value of `check_mode` - according to documentation it only accepts values `yes/no`, not `false`. 
- Replaced tags `always` with list of all tags, used in module. This makes including role much easier, as it runs only, when asked, not ie. update accidentally Prometheus, when running different tags. 
- Prepended tag names with `prometheus` to have mor unique tags, than current one.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -47,6 +47,7 @@
   delay: 2
   # run_once: true # <-- this canfalset be set due to multi-arch support
   delegate_to: localhost
+  check_mode: no
 
 - name: unpack prometheus binaries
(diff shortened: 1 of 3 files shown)


krzyzakp:
Build failed, becuase of problem getting Prometheus from sources in 2 from 3 builds. Could someone restart last build in this PR?

PR #267, opened 2020-02-15
Bump prometheus version

roidelapluie:
Note that there is a new global parameter -query_log_file- but there isn't any change needed in this role to support it.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


SuperQ:
We have a bot that deals with keeping the site up to date. It got stuck on a bad CI run, which I fixed and merged.

https://github.com/cloudalchemy/ansible-prometheus/pull/265

PR #330, opened 2021-02-01
Fix version comparison issue when using binary_local_dir

h3po:
As I was using prometheus_binary_local_dir to copy version 2.24.1, I didn't overwrite the default prometheus_version variable. For some bizarre reason, the "2.24.1" was being interpreted as an int in the version comparison in templates/prometheus.service.j2 and triggering a '<' not supported between instances of 'str' and 'int'.


h3po:
misdiagnosed, still getting the issue

PR #347, opened 2021-04-02
Fix CircleCI version bumping

SuperQ:
Better handle multiple commits and merge commits by looking backwards to
the last tagged version, rather than only the last commit.

### DIFF
--- a/bump_version.sh
+++ b/bump_version.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+#
+# Description: Generate the next release version
+
+set -uo pipefail
+
+latest_tag="$(git semver)"
+if [[ -z "${latest_tag}" ]]; then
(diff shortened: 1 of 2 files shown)


SuperQ:
This also fixes the priority order when scanning multiple log entries by picking the biggest version bump in the chain.

bump_version.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+#
+# Description: Generate the next release version
+
+set -uo pipefail
```
roidelapluie, line 5:
```suggestion
set -uneo pipefail
```

SuperQ: Nooooooope, I do not `set -e`. No silent exits.

roidelapluie (approved):

PR #349, opened 2021-04-02
Fix the circleci error in release

roidelapluie:
docker: Error response from daemon: create ~/project: "~/project"
 includes invalid characters for a local volume name, only
 "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host
 directory, use absolute path.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
           fi
       - run: |
           docker run -it --rm \
-           -v "${CIRCLE_WORKING_DIRECTORY}:/role" \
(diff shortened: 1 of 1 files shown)


SuperQ:
I fixed this in master by not using the variable.

roidelapluie:
There is another error now :(

SuperQ:
Yes, there seems to be some problems with the changelog generator. I'm looking into an option that doesn't call github API so much.

PR #359, opened 2021-05-20
Custom recording rules

ahanselka:
This PR add support for [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/). The rules are in the same format and validated in the same way as the alerting rules, so this is pretty much just the alerting rule tasks but renamed. I left the default rules blank as there aren't really generic recording rules that would serve a purpose by default (that I am aware of).

I tried to run local tests, but it seems tox may not be used anymore?


SuperQ:

PR #362, opened 2021-06-08
Changes to run multiple prometheus instances on one server

ruru4143:
## The problem

If you want a public dashboard in grafana it exposes the datasources to the public user. therefore i need to split up my datasources in public and private. For that i need multiple prometheus instances on one server

## My changes

I added 3 variables. 2 to change the system users name and group and one to change the name of the systemd service


SuperQ:

PR #363, opened 2021-06-25
fix: add family-version.yml support to variable loader task

hbjydev:
I was trying to run this from a [Rocky Linux](https://rockylinux.org) installation when I noticed that it's using `redhat.yml` to load the packages for SELinux, not `redhat-8.yml` -- this should fix that problem and allow _any_ RHEL8 derivative distro to use the correct package names, like CentOS does.


SuperQ (approved):
LGTM, Thanks!

The Lint error is unrelated.

PR #370, opened 2021-11-24
refactor(task): change checksum query method

IMSdevsecu:
Change checksum query method from lookup to uri in order to support environment proxy variable


SuperQ:

PR #375, opened 2021-12-22
Update README

sdarwin:
A small modification to the README, changing the example playbook from "- hosts: all" to "- hosts: prometheus", with an explanatory note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,11 @@ prometheus_scrape_configs:
 
 ### Playbook
 
+Typically you will create an ansible host group such as "prometheus" or "monitoring-servers", and add prometheus servers to that group. Then, run the playbook:
+ 
 ```yaml
 ---
-- hosts: all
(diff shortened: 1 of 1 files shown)


sdarwin:
This PR is low priority, not critical.  

Still, I have been thinking about it more and there are probably various ways to handle it.

- Option 1 , what was proposed here already.  

- Option 2 , instead, only make the following text change in the readme, and nothing else:
```
- hosts: prometheus
```
Quite minimal.

- Option 3 , similar as Option 1, but less verbose.  The text would be:

```
Create an ansible host group such as "prometheus" or "monitoring-servers", and add prometheus servers to that group.
```

SuperQ (approved):
LGTM, yes, I think this is a good documentation improvement. It will help with user confusion that makes them think they need Prometheus on every node.

SuperQ:

PR #376, opened 2021-12-23
Fix CircleCI

sdarwin:
It looks like the Circle CI tests have been failing for some time.   Let's check if this pull request build will succeed.

- Increase the allowed timeout for a job, with no_output_timeout: 60m
- Change the name of role, in molecule, to cloudalchemy.prometheus

### DIFF
--- a/molecule/alternative/playbook.yml
+++ b/molecule/alternative/playbook.yml
@@ -3,7 +3,7 @@
   hosts: all
   any_errors_fatal: true
   roles:
-    - ansible-prometheus
+    - cloudalchemy.prometheus
   vars:
     prometheus_binary_local_dir: '/tmp/prometheus-linux-amd64'
     prometheus_config_dir: /opt/prom/etc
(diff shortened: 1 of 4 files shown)


sdarwin:
The CI tests passed.   

There's another issue to be aware of: the entire test suite takes almost 1 hour, and occasionally more than 1 hour.   When that happens, CircleCI times out.  A possible solution could be to reduce operating system coverage, or refactor the jobs so they are more parallelized, having each task in a separate job that can run in parallel.

molecule/alternative/playbook.yml
```diff
@@ -3,7 +3,7 @@
   hosts: all
   any_errors_fatal: true
   roles:
-    - ansible-prometheus
+    - cloudalchemy.prometheus
```
SuperQ, line 6: We have pre-setup steps that create a symlink to the role name in order to handle this. But, I guess molecule now does this for us. I wonder if this is a new feature.

```
INFO     Using /home/circleci/.ansible/roles/cloudalchemy.prometheus symlink to current repository in order to enable Ansible to find the role using its expected full name.
```

sdarwin: a new feature...   And, for whatever the reason, the ansible-prometheus name wasn't being found.

SuperQ: Yea, I think the search path changed, in `cloudalchemy/ansilbe-node-exporter`, the symlink setup is different.

SuperQ (approved):
LGTM

PR #379, opened 2021-12-31
Add OracleLinux support

bekoeppel:
Oracle Linux is identified as `ansible_distribution = OracleLinux`, and needs the same package installation as CentOS 8 or RHEL.


sjpb:
This should be fixed already for all RHEL clones in `master` by aa95978755345097effca886bf116dace739d31b

PR #380, opened 2022-01-04
Retain order in yaml dictionaries

sdarwin:
In group_vars, the scrape_configs are nicely formatted with job_name at the top.  That is clear and readable.

Then, you run ansible and it generates the prometheus.yml file.  The resulting scrape configs are jumbled.  They lose the previous ordering.   

The solution is adding `sort_keys=False` to the to_nice_yaml filter.

References:
https://github.com/yaml/pyyaml/issues/110  
https://github.com/ansible/ansible/issues/76638

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -38,7 +38,7 @@
 
 - name: configure Prometheus web
   copy:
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
LGTM

PR #390, opened 2022-05-11
Allow grabbing prometheus binaries and checksums from a custom url/mirror

### LINKED ISSUE #384: Download URL
Hello,
Can you put the binaries download url in a variable.
I'm installing on a host that doesn't have internet access. Currently the download url is hardcoded. 
Thank you

gardar:
Allows grabbing prometheus binaries and checksums from a custom url/mirror, useful in closed environments where it's not possible to reach github.

I know the same could be achieved with custom tasks in a playbook and then using the `prometheus_binary_local_dir` variable, but it's pretty simple to just override the urls since they are defined anyways.

fixes #384


gardar:
Moved to: https://github.com/prometheus-community/ansible/pull/28

PR #391, opened 2022-05-11
Fix ci linting

gardar:
Fixes ansible-lint versions 5+ failing to find ansible role.

Related: https://github.com/ansible/ansible-lint/issues/1329

As well as removing deprecated mergify strict mode https://blog.mergify.com/strict-mode-deprecation/


gardar:
This should perhaps be applied from the [skeleton](https://github.com/cloudalchemy/skeleton) instead. I've created a PR there as well https://github.com/cloudalchemy/skeleton/pull/13

Although it's unclear if this repo is being synced from the skeleton, leaving this PR open in case it's not.

SuperQ:
Yes, we do sync some of these configs from skeleton.

gardar:
That's a good idea! I first noticed the issue here and I wasn't aware of the skeleton, and then once I came across the skeleton it wasn't perfectly clear where it's being used

SuperQ:

PR #405, opened 2022-12-17
Check for github action updates monthly

jlosito:
This will use dependabot to check for update against any github action which this project uses on a monthly basis. Dependabot will submit pull requests with version bumps in order to keep packages up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


SuperQ:
We already have https://github.com/cloudalchemy/auto-maintenance to do this.

